PR #15, opened 2019-08-20
Solutions for Problem1 and problem2 - Please Review

Arpan61:


Problem2.py
```diff
@@ -0,0 +1,36 @@
...
+#     def __init__(self, x):
+#         self.val = x
+#         self.left = None
+#         self.right = None
+
+class Solution:
+    def sumNumbers(self, root):
+        sm = 0
+        stack = []
+
```
sehaj-grover: try using an integer var instead of string and maintain the sum

PR #16, opened 2019-08-20
Completed Problem-1

manogna-mujje:


manogna-mujje:
> can this be done without using this much space?
> also do attempt the other problem

Completed both. Please review.

PR #17, opened 2019-08-20
Trees 2 : Tried Problem 1.

priyansh22:


priyansh22:
the code was correct, there was some issue with the IDE and return function. I changed the ide and it worked fine.

Thank you

sehaj-grover:
ok just make sure from next time send the code even if it is incorrect

PR #18, opened 2019-08-20
Trees2 done.

swaps95shah:


sehaj-grover:
try without using list to reduce space complexity

PR #22, opened 2019-09-10
Root to leaf

SaiSarathVattikuti:


sum_root_to_leaf.py
```diff
@@ -0,0 +1,31 @@
...
+	    if not root.left and not root.right:
+		    result.append(num + str(root.val))
+	    if root.left:
+		    self.dfs(root.left, result, num + str(root.val))
+	    if root.right:
+		    self.dfs(root.right, result, num + str(root.val))
+            
+            
+            #O(n)
+            #O(n)
```
kaurramanpreet: time complexity is O(N) which is correct
but space is O(Height of the tree) which will be equal to n in worst case

PR #29, opened 2019-09-11
completed

PATIL1695:
Doubt in the comment section in the code pls clarify.

### DIFF
--- a/SumRoot-leaf.py
+++ b/SumRoot-leaf.py
@@ -1,0 +1,35 @@
+"""
+The approach here is to start from the root traverse till the end of the tree where it does'nt have any
+child and at every point while traversing add the node value to the value variable. finally at the end
+we get the total.
+example - 
+
+    4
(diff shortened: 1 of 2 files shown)


SumRoot-leaf.py
```diff
@@ -0,0 +1,35 @@
...
+  9   0
+ / \
+5   1
+step 1 - 0*10 + 4 = 4
+step 2 - 4*10 + 4 = 49
+step 3 - 49*10 + 5 = 495
+step 4 - returns 495 val = 495
+step 5 - 495 + 49 *10 + 1 = 495 + 491 = 986
+step 6 returns 986
+step 7 - 986 + 4*10 + 1 = 986 + 41 = 1026
```
kaurramanpreet: yes your understanding is absolutely correct
in the end you had 0 as right child of root node so it should be 486+40!!

PR #59, opened 2019-12-30
Update inpost.java

sakshi1705:


inpost.java
```diff
@@ -0,0 +1,60 @@
+Time Complexity: O(n)
+Space Complexity: O(n) 
+
```
neha79091: time complexity will be O(n^2)

PR #73, opened 2020-01-27
Tree2 

nehabindle:
1) Problem 1 with both DF and BF solution
All the 3 codes ran successfully on leetcode

### DIFF
--- a/Solution.java
+++ b/Solution.java
@@ -1,0 +1,84 @@
+/**
+ * 
+ */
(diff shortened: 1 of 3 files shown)


Solution.java
```diff
@@ -0,0 +1,84 @@
+/**
+ * 
+ */
```
neha79091: Please write complexity, whether the solution got accepted on leetcode and if you faced any problem while solving.

PR #80, opened 2020-02-26
Completed 2 problems - Please review

PratikshaPP:


problem1.py
```diff
@@ -0,0 +1,38 @@
+# Time Complexity : O(N^2)
+# Space Complexity : O(N^2) not sure
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this : No
```
Lohith-Aswathappa: O(H) height of the tree

PratikshaPP: I realized it after yesterday's class

PR #81, opened 2020-02-26
Completed problem2 -- Please review

sayan1995:
@Lohith-Aswathappa Completed problem 2, problem 1 is an optimized solution to yesterday's class so will do it after the concept is taught today.


Lohith-Aswathappa:
good work Sayan!

PR #84, opened 2020-02-26
Problems solved afterclass

KetanKodmelwar:
@Lohith-Aswathappa

### DIFF
--- a/Problem1.java
+++ b/Problem1.java
@@ -1,0 +1,41 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No
+class Solution {
+    public TreeNode buildTree(int[] inorder, int[] postorder) {
+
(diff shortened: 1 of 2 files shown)


BhushanVasisht:
Good Work. Please try out iterative solutions too.

A recursive stack stores every argument tu pass to the recursive function. Here you are passing a HashMap in the function argument. Even if u do not get a stack overflow, when the tree has lot of nodes, your hashmap will be very large and in each recursive call the whole thing is stored in the memory stack. So u will get a Memory Limit Exceeded error. 

Only use the arguments that may potentially change inside recursive function arguments. Everything else can be kept at a global scope.

BhushanVasisht:
A recursive stack stores every argument tu pass to the recursive function. Here you are passing a HashMap in the function argument. Even if u do not get a stack overflow, when the tree has lots of nodes, your hashmap will be very large and in each recursive call the whole thing is stored in the memory stack. So u will get a Memory Limit Exceeded error.

Only use the arguments that may potentially change inside recursive function arguments. Everything else can be kept at a global scope.

PR #100, opened 2020-04-08
PR - please review

megrao:


buildTree.py
```diff
@@ -0,0 +1,32 @@
+Time complexity: O(n)
+Space complexity: O(n) for  result
```
devashishnyati, line 2: You don't generally include result in space complexity. The space will be used by the recursion stack.

buildTree.py
```diff
@@ -0,0 +1,32 @@
+Time complexity: O(n)
```
devashishnyati, line 1: Time should be O(n^2). Getting index is O(n) and you are having that operation for n elements.

devashishnyati:
Good job Megha!
Submit the other problem as well.
Verify the time and space complexity.

PR #102, opened 2020-04-08
completed problem 2

arcsoftech:


devashishnyati:
Good job Arihant!
Submit the other problem as well.

PR #103, opened 2020-04-08
adds the codes for tree2

shahakshay11:


construct_btree_from_inorder_postorder.py
```diff
@@ -0,0 +1,33 @@
+"""
+// Time Complexity : O(n)
```
devashishnyati, line 2: Shouldn't time be O(n^2)? index is O(n) operation and you are having that on n elements.

devashishnyati:
Good job Akshay!
Please verify the complexities.

construct_btree_from_inorder_postorder.py
shahakshay11: Oh, I meant to use the dict that I created to get the index. WIll update.

PR #107, opened 2020-04-10
Adding code for both problem:Please check

RutujaKaushike:


SumRootToLeaf.java
```diff
@@ -0,0 +1,32 @@
+// Time Complexity : O(m+n) m:edges, n:nodes
```
devashishnyati, line 1: So the leetcode explanation says that the time complexity is O(n^2) and in the comments people say it os O(nlogn). Copying the list is O(m) operation will be logn. This list can be the height of the tree at max. This is my understanding. Worst case it will O(n^2).

devashishnyati:
Good job Rutuja! 
Please verify the time complexity.

SumRootToLeaf.java
RutujaKaushike: I am not copying array in this solution, can you check this one?

PR #116, opened 2020-05-12
Solved probems, please review

domzhaomathematics:


tree-2.py
```diff
@@ -0,0 +1,81 @@
+# Definition for a binary tree node.
+# class TreeNode:
+#     def __init__(self, val=0, left=None, right=None):
+#         self.val = val
+#         self.left = left
+#         self.right = right
+
+#Problem 1:Construct Binary Tree from inorder and Poster order Traversal
+#Time Complexity O(n)
+#space complexity O(n)
```
sowmyadvn, line 10: Space complexity is O(H) here in worst case. We build tree, but it is the result. So, we consider only recursive stack space here for space complexity

tree-2.py
```diff
@@ -0,0 +1,81 @@
...
+
+            root.right=helper(to_split+1,right)
+            root.left=helper(left,to_split-1)
+            return root
+        return helper(0,len(inorder)-1)
+
+
+#Problem 2: Sum Root to Leaf Numbers
+#time complexity: O(n)
+#space complexity: O(h^2), h height of the tree and the strings
```
sowmyadvn, line 47: This would also be O(H) space in worst case

sowmyadvn:
Solutions are good, but please check space complexities of both the solutions.

sowmyadvn:
Solutions are good, but please check space complexities of both the solutions.

PR #124, opened 2020-05-14
Completed Design

JacobStephen9999:


iivaishnavii:
Good job, Jacob! Do submit the other solution as well.

PR #133, opened 2020-05-21
Trees-2

Siyami:


problems.js
```diff
@@ -0,0 +1,40 @@
+// ## Problem1 (https://leetcode.com/problems/construct-binary-tree-from-inorder-and-postorder-traversal/)
+
+var buildTree = function(preorder, inorder) {
+    let hash = {};
+    inorder.forEach((e, i)=>{hash[e] = i});
```
super30admin, line 9: good

problems.js
```diff
@@ -0,0 +1,40 @@
...
+    
+    let recur = function(start, end) {
+        if (start > end) return null;
+        let root = new TreeNode(preorder.shift());
```
super30admin, line 13: Shift will add O(n) time here. Better to maintain a global variable and move it to next index like class.

super30admin:
Iterative solution is missing for sum to leaf.

Please review inline comments for problem 1.

PR #135, opened 2020-06-16
SumRootToLeaf

tramatejaswini:


AbhiniveshP:
Time Complexity for both the approaches would end up O(N). 
Space Complexity for Iterative approach would be O(N) as you would store at max of N/2 nodes in the queue.
Space Complexity for Recursive approach would be O(H) where H is the height of the tree. We can say O(logN) only when we are sure that the tree is Height balanced.

AbhiniveshP:
Time Complexity for both the approaches would end up O(N). 
Space Complexity for Iterative approach would be O(N) as you would store at max of N/2 nodes in the queue.
Space Complexity for Recursive approach would be O(H) where H is the height of the tree. We can say O(logN) only when we are sure that the tree is Height balanced.

PR #152, opened 2020-06-18
Sum OF lroot to leaf number

mithi21:


AroopN:
Good work. Please verify the Time complexity. The time complexity of the approach is O(n) where n is the number of elements in the tree and not O(h). 
The space complexity if O(h) for both recursive and iterative solution.

Please try to solve other problems too.

AroopN:
Good work. Please verify the Time complexity. The time complexity of the approach is O(n) where n is the number of elements in the tree and not O(h). 
The space complexity if O(h) for both recursive and iterative solution.

Please try to solve other problems too.

PR #159, opened 2020-06-19
solved Trees-2

umamnataraj9:
solved both problems, please review


AbhiniveshP:
For both the questions, space complexity would be O(H).
Also, try iterative versions of your approach.

AbhiniveshP:
For both the questions, space complexity would be O(H).
Also, try iterative versions of your approach.

PR #171, opened 2020-06-23
Please review

snigdharreddy:


AbhiniveshP:
Nice work!!!
In building tree question, finding the index itself takes O(n) time and so overall time would end up in O(n^2).
So, use a hashmap to optimize on time complexity.

AbhiniveshP:
Nice work!!!
In building tree question, finding the index itself takes O(n) time and so overall time would end up in O(n^2).
So, use a hashmap to optimize on time complexity.

PR #181, opened 2020-07-29
Trees-2

alekhya-12:
-Solved the first problem.
-Will resubmit the PR after the second one is explained in class.
-Thanks!

### DIFF
--- a/SumToLeaf.java
+++ b/SumToLeaf.java
@@ -1,0 +1,90 @@
+/**
+Problem: Find the sum from root to leaf of the tree.
+Time Complexity: O(n), where n is the number of nodes in the tree
+Space Complexity: O(d), where d is the depth of the tree
+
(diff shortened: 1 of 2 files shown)


SumToLeaf.java
```diff
@@ -0,0 +1,90 @@
+/**
+Problem: Find the sum from root to leaf of the tree.
+Time Complexity: O(n), where n is the number of nodes in the tree
+Space Complexity: O(d), where d is the depth of the tree
```
samarthjj, line 4: O(h), where h is the height of the tree!

samarthjj:
Good work Alekya, Please also complete the second problem **Construct BST from Postorder & Inorder**

samarthjj:
Good work Alekya, Please also complete the second problem **Construct BST from Postorder & Inorder**

PR #196, opened 2020-08-03
Please Review 

firoz1905:


Sum Root to Leaf
```diff
@@ -46,7 +46,7 @@ class Solution:
                 root=root.left
             node,current_sum = stack.pop()
         
-            if root.right==None and root.left==None:
+            if root.right is None and root.left is None:
                 res += current_sum
 
         return res
```
Sumesh-H, line 52: class TreeNode:
    def __init__(self, val=0, left=None, right=None):
        self.val = val
        self.left = left
        self.right = right
class Solution:
    def sumNumbers(self, root: TreeNode) -> int:  ### Iterative Solution which is BFS
        if not root:
            return 0
        stack = []
        res = 0 # Global Variable
        current_sum=0
        while (root!= None or len(stack)!=0 ):
            while (root!= None):
                current_sum = current_sum*10 + root.val
                stack.append((root, current_sum))
                root=root.left
            root,current_sum = stack.pop(-1)

            # if root.right==None and root.left==None:
            if root.right is None and root.left is None:
                res += current_sum
            root = root.right
        return res

Sumesh-H (commented):
Good work, Firoz. Kindly check inline comments.

Sumesh-H:
Good work, Firoz. Kindly check inline comments.

PR #232, opened 2020-09-11
working on 2nd problem.

laxmikantbpandhare:


inorderPostorderTree.java
```diff
@@ -0,0 +1,71 @@
...
+        int index = -1;
+        for(int i=0;i<inorder.length;i++){
+            if(inorder[i] == root.val){
+                index = i;
+            }
+        }
+        
+        int leftSize = index;
+        int rightSize = inorder.length-1-index;
+
```
kannan-mrugapphan, line 57: Try the other approach where pointers are used to track the original arrays in recursive calls instead of creating new arrays.

kannan-mrugapphan (commented):
Good Job

kannan-mrugapphan:
Good Job

PR #255, opened 2020-10-07
trees2 completed

sujaygijre:


SumRootToLeaf.cpp
```diff
@@ -0,0 +1,41 @@
+// Time Complexity :O(N)
+// Space Complexity :O(1)
```
Lohith-Aswathappa, line 2: consider recursive stack

Lohith-Aswathappa (commented):
Good work, Sujay!
Check inline comment and try other problems as well.

Lohith-Aswathappa:
Good work, Sujay!
Check inline comment and try other problems as well.

PR #272, opened 2020-10-17
Finished Trees-2

Jaisu-1:


ramya-sruthi-d (commented):
Good job with code. But the space complexity is wrong for sumtoRoot you are using recursion there.

ramya-sruthi-d:
Good job with code. But the space complexity is wrong for sumtoRoot you are using recursion there.

PR #290, opened 2020-11-13
completed please review

Mohit0993:


samarthjj:
Good work Mohit, Please also complete the second question as well.

samarthjj:
Good work Mohit, Please also complete the second question as well.

PR #400, opened 2021-03-15
Completed Trees-2

SahanaVBhat:


nsibal (commented):
Good work!

nsibal:
Good work!

constructBinaryTree.java
```diff
@@ -0,0 +1,44 @@
+// Problem1 (https://leetcode.com/problems/construct-binary-tree-from-inorder-and-postorder-traversal/)
+// Time Complexity : O(N)
+// Space Complexity : O(N)
+public class constructBinaryTree {
```
kieraCurtis, line 4: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names should start with an upper case letter (NM_CLASS_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-should-start-with-an-upper-case-letter-nm-class-naming-convention).
Class names should be nouns, in mixed case with the first letter of each internal word capitalized. Try to keep your class names simple and descriptive. Use whole words-avoid acronyms and abbreviations (unless the abbreviation is much more widely used than the long form, such as URL or HTML).

PR #497, opened 2021-06-30
trees 2 complete please review

kushalt93:


khushal1198 (commented):
Good job with both problems Kushal!

khushal1198:
Good job with both problems Kushal!

PR #545, opened 2021-08-20
Tree-2, pls review

ysrinu1:


Rooppesh:
Good Job! Add the TC, SC and approach used!

Rooppesh:
Good Job! Add the TC, SC and approach used!

PR #647, opened 2022-01-01
solved Trees-2. Please review

akulprabhudesai:


brunda4 (commented):
SC for sumNumbers will be constant without considering the recursiveStack.

brunda4:
SC for sumNumbers will be constant without considering the recursiveStack.

PR #674, opened 2022-01-20
Completed Tree-2 "Please Review"

Abhogte:


sanjayvarma199 (commented):
Good Job!! Please check the TC for both problems

sanjayvarma199:
Good Job!! Please check the TC for both problems

PR #721, opened 2022-04-02
Completed Trees-2- Please review.

het3008:


constructtree.java
```diff
@@ -0,0 +1,75 @@
+/*
+// Time Complexity : O(n)
+// Space Complexity : O(n)
```
shrivatsav, line 3: Please specify what `n` indicates.

shrivatsav (commented):
Good Work!

shrivatsav:
Good Work!

constructtree.java
het3008: Here n is the size of the tree.

PR #796, opened 2022-06-18
Trees-2 Completed - Please Review

Ruthwik12:


Sum Root to Leaf Numbers.py
```diff
@@ -0,0 +1,25 @@
+# time complexity is o(n), where n is the size of the input
+# space complexity is o(1)
```
daredevilteja: Space complexity is wrong please check

daredevilteja (changes requested):
Please do mentioned changes

daredevilteja:
Good Work! Keep it up

Sum Root to Leaf Numbers.py
Ruthwik12: Yes, Thanks for correcting, space complexity is o(n), Recursive stack space

daredevilteja: O(h) to be specific, where h is the height of the tree

daredevilteja (commented):
Check the comment

Sum Root to Leaf Numbers.py
```diff
@@ -1,5 +1,5 @@
 # time complexity is o(n), where n is the size of the input
-# space complexity is o(1)
+# space complexity is o(n), Recursive stack space
```
daredevilteja: the recursive stack will max contain only height of the tree nodes, it will not go more than that. So here SC is O(h) not O(n)

daredevilteja (changes requested):
check the comment

Sum Root to Leaf Numbers.py
Ruthwik12: Thank you, I made changes, Thanks for bringing it up.

daredevilteja (commented):
Good work!

PR #827, opened 2022-07-03
Root to leaf sum -> DFS, global variable

ravisekhar-gonella:
https://leetcode.com/problems/sum-root-to-leaf-numbers/


Sumesh-H (commented):
Good job with both approaches. Please try the other problem as well.

Sumesh-H:
Good job with both approaches. Please try the other problem as well.

PR #879, opened 2022-08-27
Completed - Please review

mehul-0901:


Sumesh-H (commented):
Good work, add TC & SC for both the problems.

Sumesh-H:
Good work, add TC & SC for both the problems.

PR #887, opened 2022-09-11
 Completed Trees-2, please review

Suraj7696:


problem-1.py
```diff
@@ -0,0 +1,19 @@
+#Time Complexity: O(n)
+#Space Complexity: O(1)
```
daredevilteja, line 2: Sc is wrong

problem-2.py
```diff
@@ -0,0 +1,15 @@
+#Time Complexity: O(n)
+#Space Complexity: O(1)
```
daredevilteja, line 2: SC is wrong

daredevilteja (changes requested):
Please mention whether it ran successfully on leetcode or not, for every problem

daredevilteja:
Good Work! Keep it up

PR #969, opened 2022-12-18
Added 3 problems - Please Review

AmmadHassanPro:


Problem 2.java
```diff
@@ -0,0 +1,32 @@
+// Time Complexity : O(n)
+// Space Complexity : O(1)
```
daredevilteja: recheck SC

daredevilteja:
Good Work! Keep it up.

Problem 2.java
AmmadHassanPro: I fixed is now, its O(h) where h is the height of the tree

PR #1165, opened 2024-01-17
Completed Trees-2

ridashahwar:


Problem 1
```diff
@@ -0,0 +1,24 @@
+# Construct Binary Tree from Inorder and Postorder Traversal
+# Tc: O(n^2)
```
ishika2741, line 2: Submit the optimized solution as well.

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1199, opened 2024-03-22
Tree  2 done

vaishnavimore23:


ConstructTreeFromPostOrderandInorder.py
```diff
@@ -0,0 +1,24 @@
+# TC = O(n^2) --- where n is number of elements in the list
```
ishika2741, line 1: Submit it using optimized approach as well

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1240, opened 2024-07-10
Done Trees-2

RACHELSHANKAR:


Nandha2Tech (commented):
Great Work! Instead of slicing the preorder and inorder list you can pass the start and end indices to the recursive function to avoid creating new lists. This reduces TC to O(n). And also use a dictionary to store the indices of the inorder values to have O(1) search time.

Nandha2Tech:
Great Work! Instead of slicing the preorder and inorder list you can pass the start and end indices to the recursive function to avoid creating new lists. This reduces TC to O(n). And also use a dictionary to store the indices of the inorder values to have O(1) search time.

PR #1244, opened 2024-07-12
Completed

ssd9881:


Nandha2Tech (commented):
Great Work! TC for constructTreeFromIn and Post would be O(n^2) for your logic. Kindly try to do it in O(n) manner.

Nandha2Tech:
Great Work! TC for constructTreeFromIn and Post would be O(n^2) for your logic. Kindly try to do it in O(n) manner.

PR #1274, opened 2024-09-20
Trees-2 Complete

divyameharwade:


Nandha2Tech (commented):
Great Work! TC for buildTree for the logic done here would be O(n^2) (since  inorder[:ind] and inorder[ind+1:] creates new sublists). Kindly optimize it to O(n).

Nandha2Tech:
Great Work! TC for buildTree for the logic done here would be O(n^2) (since  inorder[:ind] and inorder[ind+1:] creates new sublist). Kindly optimize it to O(n).

PR #1399, opened 2025-03-13
python

divyanshi-srivastava:


problem2.py
```diff
@@ -0,0 +1,25 @@
+'''
+// Time Complexity : O(n)
+// Space Complexity : O(1)
```
ishika2741, line 3: Recursive stack space also needs to be included.

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1473, opened 2025-06-24
Done Trees-2

shreya0812:


super30admin:
Strengths:
- The use of a HashMap for O(1) lookup in the first problem is a good optimization.
- The recursive approach in the second problem is implemented correctly and efficiently.

Areas for Improvement:
- In the first problem, the variable name `idx` could be more descriptive, such as `postOrderIdx` to match the reference solution and make it clearer.
- In the second problem, the `currSum` variable could be passed directly in the helper method call without declaring it as a class member, as it is only used within the scope of the recursion.

PR #1480, opened 2025-07-07
Done Trees-2

Nav0708:


super30admin:
Strengths:
- The approach for Problem2 is correct and efficiently implemented.
- The student has a good understanding of recursion and tree traversal.

Areas for Improvement:
- For Problem1, avoid using array copying as it increases both time and space complexity. Instead, use indices to mark the boundaries of the subtrees, similar to the reference solution.
- The search for the root index in Problem1 can be optimized by using a hashmap to store the indices of the inorder array, reducing the search time from O(n) to O(1) per recursive call.
- The base case in Problem1 (inorder.length == 1) is redundant since the main base case (start > end) in the reference solution handles all cases.

PR #1484, opened 2025-07-12
PR

sindhoorpalla:


super30admin:
- The solution you provided is for a different problem (summing root-to-leaf numbers) rather than building a tree from inorder and postorder traversals. Please ensure you understand the problem requirements before coding.
- Your code quality is good, with clear structure and appropriate use of recursion. However, it's important to match the solution to the correct problem.
- For the actual problem (building a tree from inorder and postorder traversals), you would need to use a hashmap to store inorder indices and recursively construct the tree using postorder traversal.
- The time and space complexity analysis you provided is correct for the problem you solved, but it does not apply to the problem you were supposed to solve.

super30admin:
- The solution provided does not match the problem description (building a binary tree from inorder and postorder traversals). Please ensure you understand the problem requirements before starting to code.
- The code you wrote is correct and efficient for a different problem (summing root-to-leaf numbers). If you were solving that problem, your approach would be correct.
- For the actual problem, you would need to use a hashmap to store inorder indices and recursively build the tree from postorder traversal, as shown in the reference solution.
- Always double-check the problem statement and ensure your solution aligns with it.

PR #1489, opened 2025-07-21
Completed Trees-2

joelkny97:


super30admin:
- Exercise_1.py:
  - Strengths: The solution is well-commented and logically structured. The use of a hash map for O(1) lookups is efficient.
  - Areas for Improvement: Consider adding a check to ensure the input lists are not empty before proceeding, though the problem likely guarantees valid inputs.

- Exercise_2.py:
  - Strengths: The DFS approach is correctly implemented, and the use of cumulative sum is efficient. The code is easy to follow.
  - Areas for Improvement: The class variable `sum` could be avoided by returning the sum from the helper function, making the solution more functional in style.

PR #1492, opened 2025-07-23
Completed Trees-2

srilekhad:


super30admin:
- Problem1.java: The solution is correct for constructing a tree from preorder and inorder traversals, but the problem statement seems to require postorder and inorder traversals. Please ensure you are solving the correct problem. The logic and implementation are otherwise correct and efficient.
- Problem2.java: Excellent solution. The logic is clear, and the implementation is efficient. The comments are helpful and the code is well-structured.

PR #1496, opened 2025-07-29
Done Trees-2

rk-peddibhotla:


super30admin:
Strengths:
- The student has a good understanding of the problem and has implemented the solutions correctly.
- The code is well-commented and follows good practices.
- The use of a hashmap for O(1) lookup in BSTFromInAndPost is efficient.

Areas for Improvement:
- In BSTFromInAndPost, the time complexity comment should be corrected to O(n) instead of O(m).
- In SumRootToLeafNumbers, the dfs method can be simplified by moving the leaf node check outside the recursive calls to avoid redundancy.
- The TreeNode class is duplicated in both files. It would be better to have it in a separate file and import it where needed to avoid redundancy.

PR #1502, opened 2025-08-17
Janhvi Guha Trees 2

Janhvi2000:


super30admin:
- Strengths:
  - The time and space complexity analysis is accurate for both solutions.

- Areas for Improvement:
  - For ex1.py, consider using a hashmap to store the indices of the inorder traversal to improve the time complexity from O(n^2) to O(n). This would align with the reference solution and be more efficient for larger inputs.
  - While the space complexity for ex1.py is correct, it's worth noting that the slicing of lists in Python creates new lists, which can be memory-intensive. Using indices (start and end) instead of slicing could be more memory-efficient.

PR #1503, opened 2025-08-18
Trees2_Done

Sarthak15997:


super30admin:
Strengths:
- The use of a hashmap in the first problem for quick lookup is a good optimization.

Areas for Improvement:
- In the first problem, the helper function could be simplified by only passing the postorder array and the start/end indices for the inorder array, similar to the reference solution. This would reduce the number of parameters and make the code cleaner.
- In the second problem, while the iterative approach is fine, a recursive solution might be more intuitive for some readers and could be included as an alternative.

PR #1506, opened 2025-08-19
Trees-2

harsharachagiri23:


super30admin:
- Strengths:
  - The iterative approach in "Root-to-leaf.py" is efficient and well-implemented.

- Areas for Improvement:
  - For "Inorder-Postorder.py", consider using a hash map to store the indices of the inorder traversal to reduce the time complexity from O(n^2) to O(n). This would involve preprocessing the inorder list once and then using the map in the recursive calls.
  - Avoid creating new lists in each recursive call in "Inorder-Postorder.py" to save space. Instead, pass the indices of the current subarray to avoid unnecessary memory usage.
  - The space complexity comment in "Inorder-Postorder.py" could be more detailed, mentioning the worst-case scenario for a skewed tree.

PR #1518, opened 2025-09-12
Done with Trees-2

divyam2207:


super30admin:
- Problem1.py: The solution is well-implemented and follows best practices. One minor improvement could be to add a check for empty input arrays at the beginning of the `buildTree` method to handle edge cases explicitly.
- Problem2.py: The solution is correct and efficient. Consider avoiding the use of a global variable by passing the result as a parameter or using a class variable if you prefer to keep the helper function clean. This would make the code more modular and easier to test.

Overall, both solutions are well-written and solve the problems correctly. The student demonstrates a good understanding of tree traversal and recursion.

PR #1520, opened 2025-09-13
Add binary tree construction and sum algorithms

ravithakur2k:
Added ContructBinaryTree.py to build a binary tree from inorder and postorder traversals, and SumRootToLeaf.py to compute the sum of all root-to-leaf numbers in a binary tree.

### DIFF
--- a/ContructBinaryTree.py
+++ b/ContructBinaryTree.py
@@ -1,0 +1,29 @@
+# Time Complexity : O(n)
+# Space Complexity : O(n) for hashmap
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this : No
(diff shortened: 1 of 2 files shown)


super30admin:
- Strengths:
  - The student demonstrates a good understanding of tree traversal and recursion.
  - The use of a hashmap for O(1) lookups in the tree construction problem is a good optimization.
- Areas for Improvement:
  - In the SumRootToLeaf.py, the helper function returns None, which is unnecessary since the result is stored in a global variable. The function could be simplified to not return anything.
  - Adding more detailed comments explaining the recursive logic in both solutions could further enhance readability, especially for beginners.

PR #1524, opened 2025-09-14
Trees-2

v-hemish:


super30admin:
- Strengths:
  - Both solutions are correct and efficient, with clear logic and good use of recursion.
  - The comments and structure make the code easy to understand.
- Areas for Improvement:
  - In Problem1.py, consider adding a comment explaining the nonlocal idx to clarify its purpose for future readers.
  - In Problem2.py, instead of using a class variable (self.res), you could return the accumulated value from the helper function to make the code more self-contained and avoid side effects.

PR #1525, opened 2025-09-15
completed Trees-2

aditya18m:


super30admin:
- Strengths:
  - The student correctly identifies and handles edge cases.
- Areas for Improvement:
  - In problem1.cpp, the helper function could be made more concise by directly passing the postorder array and map by reference without needing to pass them repeatedly. However, this is a minor point and does not affect correctness or performance.
  - In problem2.cpp, the helper function could be simplified slightly by combining the base case and logic, but the current implementation is clear and correct.

PR #1527, opened 2025-09-17
Trees-2-Done

candida17:


super30admin:
Strengths:
- The solutions are efficient and correctly implement the required logic.

Areas for Improvement:
- For Problem1, the student could consider adding a check to ensure the input arrays are of the same length to handle edge cases more robustly.
- For Problem2, while the solution is correct, the student could mention that the space complexity is O(h) due to the recursion stack, which could be O(n) in the worst case for a skewed tree.

PR #1532, opened 2025-10-01
done

kanikanegii:


super30admin:
- Strengths:
  - The student has correctly implemented the logic for both problems.
  - The use of a helper function in both solutions is a good practice for recursive problems.
- Areas for Improvement:
  - In the SumRootToLeaf.py, the helper function could be made more concise by directly returning the sum without using a class variable (self.sum). This would make the solution more functional and reduce side effects.
  - The comments in the SumRootToLeaf.py could be more detailed to explain why the pre-order traversal is used and how it helps in accumulating the sum.

PR #1533, opened 2025-10-02
Trees-2

mdashora:


super30admin:
- The student's solution is technically correct for constructing a binary tree from preorder and inorder traversals, but it does not match the problem statement which requires postorder and inorder traversals.
- The code quality is good, with clear variable names and proper use of recursion.
- The student should ensure they understand the problem requirements before implementing a solution. The difference between preorder and postorder traversals is crucial here.
- For the correct problem (postorder and inorder), the approach would need to start from the end of the postorder array and build the right subtree before the left subtree.

PR #1538, opened 2025-11-02
Trees-2 completed

ankitakulkarnigit:


super30admin:
- The solution for `buildTree` is correct and efficient. Well done!
- The space complexity comment should be updated to O(n) to account for the hashmap storage, not just O(h) for the recursion stack.
- The inclusion of the `sumNumbers` solution is confusing and should be removed as it's unrelated to the current problem.
- The comments could be more detailed to explain why the right subtree is built before the left subtree in the `helper` function (due to postorder traversal order).

PR #1544, opened 2025-11-24
Sum Root to Leaf & Tree from Inorder & Postorder

ashishbadhe97:


super30admin:
Strengths:
- The student has provided two different approaches (void-based and int-based recursion) for the Sum Root to Leaf Numbers problem, demonstrating a good understanding of recursion.
- The student correctly handles edge cases (e.g., empty input arrays) in the tree construction problem.

Areas for Improvement:
- For the Sum Root to Leaf Numbers problem, the int-based recursion solution could be simplified further by combining the leaf node check with the base case. For example, returning `no` directly if it's a leaf node without separate checks.
- In the tree construction problem, the variable names could be more descriptive (e.g., `rootIdx` could be `postOrderIdx` to match the reference solution and improve clarity).

PR #1545, opened 2025-12-03
Solved Trees 2 problems

aahiawasthee:


super30admin:
Strengths:
- The student has correctly implemented both solutions with optimal time and space complexity.
- The use of a hashmap for quick lookups in the binary tree construction is a good optimization.

Areas for Improvement:
- In the sum-root-to-leaf solution, the `result` variable could be passed as a parameter to the helper function to make the solution more functional and avoid using a class-level variable. This would make the code more modular and easier to test.
- The comment "we do right tree first cause postorder is right root left" in the binary tree construction is slightly misleading. It would be clearer to say that we process the right subtree first because in postorder, the root is at the end, and the right subtree is just before the root.

PR #1546, opened 2025-12-08
Day17- Construct-binary-tree-from-inorder-and-postorder-traversal, Sum-root-to-leaf-numbers

tanvipandhre:
Day17- Construct-binary-tree-from-inorder-and-postorder-traversal, Sum-root-to-leaf-numbers

### DIFF
--- a/Construct-binary-tree-from-inorder-and-postorder-traversal.java
+++ b/Construct-binary-tree-from-inorder-and-postorder-traversal.java
@@ -1,0 +1,42 @@
+/**
+ * Definition for a binary tree node.
+ * public class TreeNode {
+ *     int val;
+ *     TreeNode left;
+ *     TreeNode right;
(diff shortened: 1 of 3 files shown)


super30admin:
- Strengths:
  - The logic for both problems is correct and well-implemented.
  - The code is readable and well-structured.
  - The README.md provides a clear explanation of the approach.

- Areas for Improvement:
  - For the buildTree problem, consider using a hashmap to store the indices of the inorder array. This will improve the time complexity from O(n^2) to O(n).
  - Avoid creating new arrays for each recursive call in the buildTree problem. Instead, use indices to represent the subarrays, which will reduce the space complexity.
  - The sumNumbers solution is efficient, but you could consider using an iterative approach (e.g., using a stack) to avoid the recursion stack overhead, though this is optional.

PR #1550, opened 2025-12-21
Trees-2-Completed

PrasiddhShah:


super30admin:
- Strengths:
  - The code is clean, readable, and well-commented.
  - The solution for problem2.java is optimal and correctly implemented.
  - Good use of recursion and base cases.

- Areas for Improvement:
  - For problem1.java, ensure you understand the problem requirements correctly. The problem seems to require building a tree from postorder and inorder traversals, not preorder and inorder.
  - Avoid array copying in recursive calls to improve time and space complexity. Instead, use indices to mark the current subarray range, similar to the reference solution.
  - Consider using a hashmap to store the indices of the inorder array to reduce the time complexity from O(n^2) to O(n).
  - Be accurate in stating space complexity, especially when recursive stack space is involved.

PR #1562, opened 2026-01-21
Trees-2

praxpk:


super30admin:
- Strengths:
  - The student has correctly implemented the solution with the right time and space complexity.
  - The use of a helper function to manage recursion is a good practice.
- Areas for Improvement:
  - The student included an unrelated file (`sum_of_root_leaf.py`) in the submission, which should be avoided unless it's part of the problem set.
  - While the comments are helpful, they could be more detailed to explain the logic behind the recursion and the use of the hashmap.

PR #1563, opened 2026-01-22
Solutions for Trees-2 problems

neeeraj1999:


super30admin:
- Strengths:
  - Both solutions are correct and efficient, matching the reference solutions in time and space complexity.
  - The use of helper functions and descriptive variable names enhances readability.

- Areas for Improvement:
  - In Problem1, consider adding a brief comment explaining why the right subtree is built before the left subtree, as this might not be immediately obvious to someone reading the code.
  - In Problem2, while the solution is correct, it could be slightly optimized by returning the sum directly from the DFS function instead of using a nonlocal variable. This would make the code slightly cleaner.

PR #1566, opened 2026-01-30
Adding Trees-2 solutions

subbu4061:


super30admin:
Strengths:
- The solutions are correct and efficient for both problems.
- The code is well-organized and readable with meaningful variable names.
- The student provided explanations for both solutions, showing good communication.

Areas for Improvement:
- For the tree construction problem, note that the recursion order is important: building the right subtree first because the postorder array is processed from the end (which gives the root, then right, then left). This is correctly implemented, but the student should ensure they understand why.
- In the second problem (Sum Root to Leaf Numbers), the first solution (without global variable) is generally preferred as it avoids instance variables and is more thread-safe. The second solution with a global variable might be acceptable in a coding interview if explained, but the first is cleaner.
- The student should be consistent with code formatting (e.g., spacing and indentation). For example, in the tree construction code, there is an extra space in `inleft, inright` in the helper call. Also, in the second solution for Sum Root to Leaf Numbers, the helper method has inconsistent spacing (e.g., `helper (TreeNode root, int curr)` has a space before the parenthesis). This is minor but worth noting for clarity.

Overall, the student has done an excellent job.

PR #1570, opened 2026-02-09
Complete Trees-2

paridhimalviya:
Problem1 (https://leetcode.com/problems/construct-binary-tree-from-inorder-and-postorder-traversal/)

Problem2 (https://leetcode.com/problems/sum-root-to-leaf-numbers/)

### DIFF
--- a/ConstructBTFromInorderPostorder.swift
+++ b/ConstructBTFromInorderPostorder.swift
@@ -1,0 +1,52 @@
+//
+//  ConstructBinaryTreeFromInorderTraversal.swift
+//  DSA-Practice
+//
+//  Created by Paridhi Malviya on 1/19/26.
+//
+
+/**
+ * Definition for a binary tree node.
+ * public class TreeNode {
+ *     public var val: Int
+ *     public var left: TreeNode?
+ *     public var right: TreeNode?
+ *     public init() { self.val = 0; self.left = nil; self.right = nil; }
+ *     public init(_ val: Int) { self.val = val; self.left = nil; self.right = nil; }
+ *     public init(_ val: Int, _ left: TreeNode?, _ right: TreeNode?) {
+ *         self.val = val
+ *         self.left = left
+ *         self.right = right
+ *     }
+ * }
+ */
+class BinaryTreeFromInorderAndPostorder {
+    var postOrderIdx = 0
+    func buildTree(_ inorder: [Int], _ postorder: [Int]) -> TreeNode? {
+        postOrderIdx = postorder.count - 1
+        var inMap = Dictionary<Int, Int>()
+        for (index, element) in inorder.enumerated() {
+            inMap[element] = index
+        }
+        return constructTree(postorder, inMap: inMap, inStart: 0, inEnd: inorder.count - 1)
+    }
+
+    func constructTree(_ postorder: [Int], inMap: Dictionary<Int, Int>, inStart: Int, inEnd: Int) -> TreeNode? {
+        //base
+        if (inStart > inEnd) {
+            return nil
+        }
+        
+        //logic
+        let rootVal = postorder[postOrderIdx]
+        postOrderIdx -= 1
+        let root = TreeNode(val: rootVal, left: nil, right: nil)
+        let rootIdx = inMap[rootVal]!
+
+        //recurse
+        root.right = constructTree(postorder, inMap: inMap, inStart: rootIdx + 1, inEnd: inEnd)
+        root.left = constructTree(postorder, inMap: inMap, inStart: inStart, inEnd: rootIdx - 1)
+
+        return root
+    }
+}
--- a/RootToLeafSum.swift
+++ b/RootToLeafSum.swift
@@ -1,0 +1,80 @@
+//
+//  RootToLeafSum.swift
+//  DSA-Practice
+//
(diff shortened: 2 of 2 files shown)


super30admin:
Strengths:
- You correctly identified the approach of using a hash map to store inorder indices for O(1) lookups.
- The recursive function correctly handles the construction of the tree by first building the right subtree and then the left subtree, which is essential for postorder traversal.
- The base case for the recursion is correctly implemented.

Areas for improvement:
- The `postOrderIdx` should not be a class property because it will be shared across multiple calls to `buildTree`. If the same instance is used to build multiple trees, the index will not reset correctly. Instead, you should pass the index by reference (e.g., using an inout parameter) or restructure the code to avoid state. Alternatively, you can make `postOrderIdx` a local variable and pass it by reference to the helper function.

- The helper function `constructTree` should be marked as private to avoid exposing it as part of the public API.

- The class name `BinaryTreeFromInorderAndPostorder` is descriptive but a bit verbose. Consider shortening it to something like `Solution` if that aligns with the problem context.

- The submission includes an unrelated file `RootToLeafSum.swift`. For the purpose of this problem, only the relevant code should be submitted. Ensure that your solution is focused on the problem at hand.

- In Swift, it's common to use lowerCamelCase for parameter names. The parameter `inMap` should be `inMap` (which it is), but `inStart` and `inEnd` should be `inStart` and `inEnd` to follow Swift naming conventions.

Revised Solution Suggestion:
To fix the issue with `postOrderIdx`, you can make it a local variable in `buildTree` and pass it as an inout parameter to the helper function. Alternatively, you can define the helper function inside `buildTree` to capture the index. Here's a revised version:

```swift
class Solution {
    func buildTree(_ inorder: [Int], _ postorder: [Int]) -> TreeNode? {
        var inMap = [Int: Int]()
        for (index, value) in inorder.enumerated() {
            inMap[value] = index
        }
        var postIndex = postorder.count - 1
        return buildTreeHelper(postorder, &postIndex, inMap, 0, inorder.count - 1)
    }
    
    private func buildTreeHelper(_ postorder: [Int], _ index: inout Int, _ inMap: [Int: Int], _ inStart: Int, _ inEnd: Int) -> TreeNode? {
        if inStart > inEnd { return nil }
        let rootVal = postorder[index]
        index -= 1
        let root = TreeNode(rootVal)
        let rootIndex = inMap[rootVal]!
        root.right = buildTreeHelper(postorder, &index, inMap, rootIndex + 1, inEnd)
        root.left = buildTreeHelper(postorder, &index, inMap, inStart, rootIndex - 1)
        return root
    }
}
```

super30admin:
Your solution is on the right track and demonstrates a good understanding of the problem. You correctly used a hash map to store inorder indices for efficient root location, and you correctly traversed the postorder array from the end to the beginning. The recursion order (right subtree first, then left) is correct for postorder traversal.

However, there are a few issues to address:
1. The TreeNode initializer you used does not match the provided definition. The provided TreeNode class does not have an initializer with named parameters. You should use `TreeNode(rootVal)` for the node without children, or `TreeNode(rootVal, left, right)` when specifying children. In your code, you are creating a node with no children, so you can simply use `TreeNode(rootVal)`.

2. The class-level variable `postOrderIdx` is reset in `buildTree`, which is good. However, consider that if the same instance is used to build multiple trees, the variable might be modified concurrently. While the problem does not require thread safety, it is generally better to avoid statefulness when possible. You could make the variable local to the method and pass it by reference (e.g., as an inout parameter) in Swift, or use a nested function that captures a variable. For example, you could define a local variable `var postOrderIdx = postorder.count - 1` in `buildTree` and then have a nested function that uses and updates it.

3. You included an unrelated file (`RootToLeafSum.swift`) in your submission. This file is not relevant to the current problem and should be removed to keep the solution focused.

4. The function `constructTree` is well-named, but the parameters `inStart` and `inEnd` are clear. However, consider using more descriptive names like `inorderStart` and `inorderEnd` for clarity.

To fix the TreeNode initialization, change:
```swift
let root = TreeNode(val: rootVal, left: nil, right: nil)
```
to:
```swift
let root = TreeNode(rootVal)
```

Alternatively, if you want to explicitly set left and right to nil, you can use:
```swift
let root = TreeNode(rootVal, nil, nil)
```
But the first option is simpler.

Here is a revised version of your solution without the stateful variable (using a nested function):

```swift
class Solution {
    func buildTree(_ inorder: [Int], _ postorder: [Int]) -> TreeNode? {
        var inMap = [Int: Int]()
        for (index, value) in inorder.enumerated() {
            inMap[value] = index
        }
        var postOrderIdx = postorder.count - 1
        func constructTree(_ inStart: Int, _ inEnd: Int) -> TreeNode? {
            if inStart > inEnd { return nil }
            let rootVal = postorder[postOrderIdx]
            guard let rootIdx = inMap[rootVal] else { return nil }
            postOrderIdx -= 1
            let root = TreeNode(rootVal)
            root.right = constructTree(rootIdx + 1, inEnd)
            root.left = constructTree(inStart, rootIdx - 1)
            return root
        }
        return constructTree(0, inorder.count - 1)
    }
}
```

This avoids the class-level variable and uses a nested function to capture `postOrderIdx` and `inMap`.

Overall, your solution is almost correct but for the compilation error. With that fixed, it would work.

PR #1574, opened 2026-02-27
[Sum number && build post order and inorder]

takchiks:


super30admin:
Strengths:
- The student uses a hashmap to store inorder indices, which is efficient.
- The recursive approach is correctly implemented with global index tracking for postorder.
- The order of building the tree (right subtree first then left) is correct for postorder.

Areas for improvement:
1. The code includes an unrelated method `sumNumbers` which is not part of the problem. This should be removed to keep the solution focused and avoid confusion.
2. The variable names could be improved for clarity: `postorderIdx` is acceptable, but the parameters in `helperTree` are `left` and `right` which represent the start and end indices in the inorder array. It might be better to name them `inStart` and `inEnd` to avoid confusion with the postorder array.
3. The initial call to `helperTree` uses `postorderIdx` as the right boundary. While it is correct because the arrays have the same length, it would be clearer to use `inorder.length-1` directly to make it explicit that the boundaries are for the inorder array.
4. The code lacks comments explaining the approach. Adding comments would make it more understandable.

Suggested corrections:
- Remove the `sumNumbers` method.
- Change the initial call to: `return helperTree(postorder, 0, inorder.length-1);`
- Consider renaming parameters in `helperTree` to `inStart` and `inEnd`.

PR #1580, opened 2026-03-15
Add implementation for leetcode problems 106, 129

rishigoswamy:


super30admin:
### Construct Binary Tree from Inorder and Postorder Traversal (leetcode_106.py)
Your solution has a good approach by using a hashmap for O(1) index lookup and a global index to traverse the postorder array. However, there is a minor issue in the initial call to `createTree`: you are passing `len(postorder)-1` as the `right` parameter, which represents the end index in the inorder array. Since the lengths of inorder and postorder are the same, this is correct, but for clarity, you should use `len(inorder)-1` instead. This makes the code more readable and less error-prone if the arrays were to have different lengths (even though the problem states they are the same).

Also, note that the parameters `left` and `right` in `createTree` are indices in the inorder array, not the postorder array. Your recursive calls are correctly using these indices to split the inorder array.

Your code is efficient with O(n) time and space complexity. The use of a global index for postorder traversal is correct and efficient.

You also provided two alternative implementations in comments. These are less efficient: the first one uses array slicing which creates new arrays and has O(n) time per slice, leading to O(n^2) time complexity. The second one uses `inorder.index` which is O(n) per call, also leading to O(n^2) time. It's good that you recognized the efficient method and implemented it.

To improve:
- Change the initial call to `return self.createTree(postorder, 0, len(inorder)-1)` for clarity.
- Consider renaming the parameters in `createTree` to `in_left` and `in_right` to make it clear they are indices for the inorder array.
- Remove the commented code to keep the solution clean and focused.

Overall, your solution is correct and efficient. With a small change for clarity, it would be excellent.

VERDICT: PASS

---

### Root to Leaf Sum (leetcode_129.py)
Your solution is correct in logic and follows a BFS approach, which is a valid way to solve the problem. The code is clear and well-commented, which is good practice. However, there is a performance issue with using `list.pop(0)` in Python, as it is not efficient for large lists. Instead, you should use `collections.deque` for efficient popping from the front. This would improve the time complexity to O(n) for the entire traversal.

Here's a suggested improvement for your code:
- Replace the list with `collections.deque` to achieve O(1) for popping from the front.
- Initialize the deque with `from collections import deque` and then use `queue = deque([[root, root.val]])` and `node, currval = queue.popleft()`.

Additionally, you should consider handling the case where the root is None to avoid potential errors. Although the problem states the number of nodes is at least 1, it's good practice to check for edge cases.

Overall, your solution is correct but can be optimized for time by using a more efficient data structure for the queue.

VERDICT: NEEDS_IMPROVEMENT

PR #1588, opened 2026-05-04
Adding solution-2

nikhylw:
Solution 2 - W4_2_sum_root_to_leaf.py


super30admin:
### Construct Binary Tree from Inorder and Postorder Traversal (W4_2_sum_root_to_leaf.py)
It seems there has been a mix-up. The problem you were asked to solve is "Construct Binary Tree from Inorder and Postorder Traversal", but your solution is for "Sum Root to Leaf Numbers". Please double-check the problem statement.

For the correct problem, you need to build a binary tree from its inorder and postorder traversal arrays. The key idea is to use the last element in the postorder array as the root, then locate this root in the inorder array to split the array into left and right subtrees. You can then recursively build the left and right subtrees.

Here are some tips for solving the correct problem:
1. Use a hash map to store the indices of the inorder array for quick root position lookup.
2. Use a recursive function that takes indices for the current segment of the inorder array.
3. Note that in postorder, the root is at the end, and you should process the right subtree before the left subtree when building recursively (because the postorder array is structured as [left, right, root], so when traversing backwards, you get root, then right, then left).

VERDICT: NEEDS_IMPROVEMENT

---

### Root to Leaf Sum
Your solution is on the right track and you have correctly implemented the DFS traversal. However, there is a subtle issue with the order of operations. You are checking for leaf node after recursively processing both children. This is inefficient because you are making recursive calls for null children (which immediately return) and then checking the leaf condition. The standard approach is to check for leaf node immediately after updating the current number, and if it is a leaf, add to the result and return without going into children (since they are null). This way, you avoid unnecessary recursive calls for null children.

Here is how you can improve your helper function:
```
def helper(self, root, curr_sum):
    if root is None:
        return
    
    curr_sum = curr_sum * 10 + root.val
    
    # Check if it's a leaf node
    if root.left is None and root.right is None:
        self.result += curr_sum
        return  # No need to go further
    
    # Otherwise, recurse to children
    self.helper(root.left, curr_sum)
    self.helper(root.right, curr_sum)
```

This change will make your code more efficient by avoiding unnecessary recursive calls for null children. Also, it is more intuitive because you handle the leaf node immediately when you encounter it.

Your time and space complexity analysis is correct. Keep up the good work!

VERDICT: NEEDS_IMPROVEMENT

